PR #119, opened 2026-04-09
Add greeting folder plugin with time-bucket cache (Plan 40)

spideyz0r:
## Summary
- Migrates `plugins/greeting.py` to a folder-based manifest plugin (`plugins/greeting/`) with `plugin.yaml` self-registering the route — `greeting:` removed from `routes.yaml`
- Adds `_cache_key()` that buckets by time-of-day (`morning/afternoon/evening/night`) with 1h TTL and 90min max-stale, so "bom dia" / "boa noite" are instant from cache
- Supports `refresh_only=True` for background warmup via `cache_auto_refresh`; cache read/write failures log WARNING and fall through to live generation

## Test plan
- [x] 13 new tests in `tests/test_greeting_plugin.py` covering all cache paths (hit, miss, stale, read failure, write failure, no cache, refresh_only)
- [x] Full suite: 860 passed
- [x] Add `greeting` entry to `cache_auto_refresh` in `~/.sandvoice/config.yaml`


spideyz0r:

### 1. `_cache_key()` uses machine local TZ instead of `config.timezone` (plugin.py line 22)
- ✅ Added `_resolve_tz(config)` helper that mirrors the `TaskScheduler._resolve_tz()` pattern from `common/scheduler.py`: tries `ZoneInfo(config.timezone)`, warns and returns `None` on failure, and short-circuits when `ZoneInfo` is unavailable (Python < 3.9)
- ✅ `_cache_key()` now calls `datetime.now(tz)` with the resolved zone (or `None` for local time fallback)
- ✅ Added `test_valid_timezone_used_when_available` — verifies no exception when `config.timezone = "UTC"` and returns a valid bucket
- ✅ Added `test_invalid_timezone_falls_back_to_local` — verifies a WARNING is logged and a valid bucket is still returned for `"INVALID/TIMEZONE"`

**Test Results:**
- Total tests: 862
- All passing ✅

spideyz0r:

### 1. `_cache_key` signature breaks `_derive_cache_key` warmup (plugin.py line 32)
- ✅ Dropped unused `route` parameter from `_cache_key`; new signature is `_cache_key(config=None)` so it can be called with no args without raising `TypeError`
- ✅ Added `elif param == "config": kwargs["config"] = config` to `_derive_cache_key()` in `sandvoice.py` — the greeting warmup now receives `config` and can correctly resolve the timezone bucket at startup

### 2. `test_invalid_timezone_falls_back_to_local` fails on Python < 3.9 (tests line 76)
- ✅ Added `skipTest` guard (same pattern as `test_valid_timezone_used_when_available`) — on Python < 3.9 `ZoneInfo = None` and the plugin falls back silently without a WARNING, so the `assertLogs` would fail; skipping is correct

### 3. `_make_cache_entry` helper is unused (tests line 18)
- ✅ Removed the unused helper

**Test Results:**
- Total tests: 862
- All passing ✅

spideyz0r:

### 1. `process()` live generation path not wrapped in try/except (plugin.py line 49)
- ✅ Wrapped the full live generation block (weather route, weather response, LLM call, cache write) in `try/except Exception`
- ✅ On failure: logs `ERROR`, returns `None` when `refresh_only=True` (background warmup — no TTS anyway), returns `"Unable to greet you right now. Please try again."` otherwise
- ✅ Added `test_llm_error_returns_friendly_message` — verifies ERROR is logged and a non-empty string is returned
- ✅ Added `test_llm_error_with_refresh_only_returns_none` — verifies ERROR is logged and `None` is returned for background refresh

### 2. Plan doc shows outdated `_cache_key(route, config)` signature (plan doc line 47)
- ✅ Updated the code snippet to `_cache_key(config=None)` with `_resolve_tz(config)` call
- ✅ Updated the implementation note on line 92 from `_cache_key(route, config)` to `_cache_key(config=None)` with timezone note

**Test Results:**
- Total tests: 864
- All passing ✅

spideyz0r:

### 1. `_resolve_tz()` warns on non-string config.timezone (plugin.py line 26)
- ✅ Added `isinstance(tz_name, str)` guard before attempting `ZoneInfo()` resolution — non-string values (MagicMock, None, int) now silently fall back to local time without emitting a WARNING

### 2. `_key_for_hour()` test helper triggers spurious warnings (tests line 18)
- ✅ Set `config.timezone = None` explicitly in the helper so `_resolve_tz()` short-circuits cleanly instead of attempting resolution on a truthy MagicMock attribute

### 3. Plan doc "Behaviour" section inaccurate about stale entries (plan doc line 61)
- ✅ Split "Cache hit" into fresh and stale-but-acceptable cases, clarifying that entries within `max_stale_s` are served from cache (no regeneration)
- ✅ Corrected "up to `ttl_s` seconds stale" → "up to `max_stale_s` seconds stale" in the trade-off note

**Test Results:**
- Total tests: 864
- All passing ✅

spideyz0r:

### 1. `refresh_only` not strictly checked (plugin.py line 44)
- ✅ Changed to `route.get('refresh_only') is True` — only boolean `True` triggers refresh-only mode; truthy non-bool values like `'false'` or `1` no longer bypass cache reads

### 2. Exception handler discards stack trace (plugin.py line 86)
- ✅ Added `exc_info=logger.isEnabledFor(logging.DEBUG)` to the error log — mirrors the pattern used in `plugins/weather/plugin.py`; traceback is available at debug level without polluting normal log output

### 3. `test_valid_timezone_used_when_available` didn't verify TZ was used (tests line 58)
- ✅ Replaced the loose `assertIn(result, [...])` assertion with `datetime.now.assert_called_once_with(ZoneInfo("UTC"))` — now actually verifies that the correct ZoneInfo is passed to `datetime.now()`

### 4. `test_refresh_only_skips_cache_read` missing cache assertion (tests line 178)
- ✅ Added `cache.get.assert_not_called()` to verify the cache read is bypassed, not just that the result is `None`

**Test Results:**
- Total tests: 864
- All passing ✅

spideyz0r:

### 1 & 2. Duplicate scheduler tasks from time-dependent cache key (sandvoice.py line 85, plugin.py line 40)
- ✅ Changed `task_name = f"cache_refresh:{cache_key}"` → `task_name = f"cache_refresh:{plugin_name_raw}:{query}"` in `sandvoice.py`
- The task name is now derived from stable config (plugin name + query string) rather than the computed cache key, so the same task name is used on every restart regardless of time-of-day — no duplicate tasks accumulate across restarts

### 3. `_resolve_tz()` logs WARNING on every call with invalid timezone (plugin.py line 25)
- ✅ Added `_TZ_CACHE` module-level dict: each `tz_name` is resolved at most once per process lifetime; subsequent calls return the cached `ZoneInfo` (or `None`) directly without calling `ZoneInfo()` or re-logging
- ✅ Updated `test_invalid_timezone_falls_back_to_local` to pop `"INVALID/TIMEZONE"` from `_TZ_CACHE` before the assertion so the warning path is always exercised regardless of test execution order

**Test Results:**
- Total tests: 864
- All passing ✅

spideyz0r:
## Round 7 Copilot Feedback

### Stale old-format tasks on upgrade (sandvoice.py line 515) — Skipped
- ⏭️ **Reason**: Copilot marked this as low confidence. Old `cache_refresh:{cache_key}` tasks that remain in the DB after the naming change to `cache_refresh:{plugin}:{query}` would only run `refresh_only=True` — extra cache refreshes, no correctness impact.

  A migration to purge old-format tasks would require reverse-engineering the previous cache-key shape per plugin (e.g. `news:https://...`, `weather:[loc, unit]`, `greeting:morning`) or adding a name-prefix query to `SchedulerDB` that doesn't exist today. For a personal project where the operator controls the DB, `rm ~/.sandvoice/scheduler.db` is the practical reset path. The complexity is not justified.

No code changes this round.

**Test Results:**
- Total tests: 864
- All passing ✅

PR #124, opened 2026-04-11
Add OpenAI provider implementations (Plan 42)

spideyz0r:
## Summary

- `OpenAILLMProvider` — generate_response, stream_response_deltas, define_route, text_summary
- `OpenAITTSProvider` — text_to_speech with chunking, retry, and cleanup on failure
- `OpenAISTTProvider` — transcribe (rename of transcribe_and_translate; all task modes preserved)
- 35 tests, 91% coverage across all three providers
- `AI` class unchanged — providers coexist alongside it

## Planning Document
Implements plan/completed/42-openai-provider-implementations.md

## Key design decisions
- `conversation_history` is a parameter to LLM methods — providers do not own or mutate it
- `_build_system_role()` helper eliminates the duplication between `generate_response` and `stream_response_deltas`
- `stream_response_deltas` has no retry decorator (streaming retry semantics are ambiguous)
- `STTProvider.transcribe` uses `audio_file_path=None` to fall back to configured tmp path

## Testing
- [x] All 916 tests pass (existing suite unchanged)
- [x] Coverage >80% for all three new files (91% combined)
- [x] Tested on Mac M1
- [ ] Tested on Pi 3B

## Notes
Plan 43 (AI facade migration) wires these providers into `AI` and removes the duplicate logic.


spideyz0r:

### 1. `generate_response` missing `user_input` in messages (`openai_llm.py` line 72)
- ✅ Added `{"role": "user", "content": user_input}` as the final element of the messages list. The provider is now self-contained — the current prompt is always included regardless of what the caller passes in `conversation_history`.

### 2. `stream_response_deltas` missing `user_input` in messages (`openai_llm.py` line 120)
- ✅ Same fix applied. Final message is now the current `user_input`.

### 3. `text_summary` `extra_info` not interpolated — missing `f` prefix (`openai_llm.py` line 204)
- ✅ Fixed: `f"Consider that this is the question of the user: {extra_info}\n\n"` — added `f` prefix and a newline separator.

### 4. Typo "limite" → "limit" in summary prompt (`openai_llm.py` line 200)
- ✅ Fixed.

### 5. `retry_with_backoff` can't see `self._config` — needs `self.config` (`openai_tts.py` line 17)
- ✅ Added `self.config = config` in `OpenAITTSProvider.__init__`. The decorator uses `getattr(args[0], 'config', None)` so the alias is required.

### 6. Same `self.config` issue in `OpenAISTTProvider` (`openai_stt.py` line 14)
- ✅ Added `self.config = config` in `OpenAISTTProvider.__init__`.

### 7. Double error handling in `openai_stt.py` — translation exception re-caught as "OpenAI Whisper" (`openai_stt.py` line 82)
- ✅ Restructured into two separate try blocks: the first covers file open + audio transcription/translation via Whisper; the second (outside the first) covers the chat-completions translation step. Errors from each block are now correctly labelled without double-handling.

### 8. Test missing assertion that `user_input` is the final message (`test_openai_providers.py` line 79)
- ✅ Added assertions: `messages[-1]["role"] == "user"` and `messages[-1]["content"] == "hi"`.

### 9. Same `self.config` issue in `OpenAILLMProvider` (`openai_llm.py` line 20)
- ✅ Added `self.config = config` in `OpenAILLMProvider.__init__`.

**Test Results:** 916 passed, 0 failed ✅

spideyz0r:

### 1. Retry decorator was ineffective on generate_response/define_route/text_summary (multiple lines)
- ✅ Split each into a decorated private `_call_*` method (raises on failure, retried) and a public method (catches after retries exhaust). Mirrors the existing `_generate_tts_files`/`text_to_speech` pattern from the TTS provider.

### 2. user_input was not included in messages sent to the model (stream_response_deltas / generate_response)
- ✅ Added `_build_messages()` helper that appends `user_input` (prefixed with `"User: "` to match conversation_history format) as the final message in the messages list.

### 3. Debug log included response content instead of metadata-only
- ✅ Replaced `generate_response` debug log with a metadata-only log (length, verbosity, has_extra_info, model, history size). No response text in logs.

### 4. Unused import os in openai_llm.py
- ✅ Removed.

### 5. Retry split pattern for define_route / text_summary (follow-on from #1)
- ✅ Both methods now follow the same `_call_*` / public split as generate_response.

### 6. _make_config() temp dir leak — nothing cleaned it up
- ✅ Changed `_make_config()` to use `/tmp` as default `tmp_files_path` (no directory created). Added `TemporaryDirectory` setUp/tearDown to `TestOpenAITTSProviderTextToSpeech`, which is the only class that actually writes files to disk.

### 7. Unused imports threading and SimpleNamespace in test file
- ✅ Removed both.

### 8. Empty tearDown() in TestOpenAISTTProviderTranscribe
- ✅ Removed.

### 9. f-string missing in text_summary extra_info interpolation (addressed in previous round)
- ✅ Already fixed in round 1. Confirmed still correct.

**Test Results:**
- Total tests: 916
- All passing ✅

spideyz0r:

### 1. Grammar errors in `_call_text_summary()` system prompt (line 224)
- ✅ Rewrote all lines: "You're a bot summaries texts" → "You are a bot that summarizes texts", "must content" → "must contain", "will be in json format" → "must be in JSON format", and other clarity improvements. Better wording reduces prompt ambiguity and improves model compliance.

### 2. `warnings.filterwarnings()` mutating global filter on every call (line 37, openai_tts.py)
- ✅ Replaced the global `warnings.filterwarnings("ignore", category=DeprecationWarning)` with a `with warnings.catch_warnings(): warnings.simplefilter(...)` context scoped to the `response.stream_to_file()` call — the actual source of the deprecation warning. No global side effects.

### 3. `_build_messages()` could duplicate user_input if history already contains it (line 67) — Skipped
- ⏭️ Reason: The method's docstring explicitly documents the contract — `conversation_history` contains prior turns only, and `user_input` is the current prompt. Adding a de-duplication guard would obscure the contract and silently accommodate incorrect callers. The design is intentional and clear.

### 4. `transcribe()` prints error messages to stdout (line 62, openai_stt.py) — Skipped
- ⏭️ Reason: `docs/PATTERNS.md` explicitly states: "User-facing output (always shown, not gated) uses `print()` directly — this is intentional and correct in any module (including `common/`) when the output is part of the interactive UX." The `print(error_msg)` calls follow this pattern consistently across the codebase.

**Test Results:**
- Total tests: 916
- All passing ✅

spideyz0r:

### 1–3. Dual config storage (self._config + self.config) in all three providers
- ✅ Removed `self._config = config` from `OpenAILLMProvider`, `OpenAITTSProvider`, and `OpenAISTTProvider`. All internal references updated to `self.config`. Single attribute eliminates ambiguity and still satisfies `retry_with_backoff`'s `self.config` lookup.

**Test Results:**
- Total tests: 916
- All passing ✅

spideyz0r:

### 1. warnings.catch_warnings() recreated on every chunk iteration (line 58, openai_tts.py)
- ✅ Moved `with warnings.catch_warnings(): warnings.simplefilter("ignore", DeprecationWarning)` to wrap the entire chunk loop. The filter is now set once per request instead of per chunk — less overhead and clearer suppression scope.

**Test Results:**
- Total tests: 916
- All passing ✅

spideyz0r:

### 1. transcript.text accessed outside Whisper try block (line 64, openai_stt.py)
- ✅ Moved `source_text = transcript.text or ""` inside the first try block (after the `with open` context, before the except clauses). An `AttributeError` from an unexpected API response shape is now caught and labeled as an "OpenAI Whisper" error with proper logging, consistent with all other Whisper errors. The two-try-block structure (Whisper vs chat-completions) is preserved.

**Test Results:**
- Total tests: 35 provider tests
- All passing ✅

PR #125, opened 2026-04-12
Migrate AI to provider facade (Plan 43)

spideyz0r:
## Summary

- `AI.__init__` now takes `(llm, tts, stt, config)` — all capability calls delegate to provider instances
- `AI.from_config(config)` is the standard construction path; `sandvoice.py` updated at all three call sites
- `_build_llm/tts/stt_provider()` factory helpers raise `ValueError` on unknown provider name (fast startup fail)
- `llm_provider`, `tts_provider`, `stt_provider` config keys added (all default to `"openai"`)
- `AI` owns `conversation_history`; providers receive snapshots and never mutate it
- Deferred imports in `_build_*_provider` helpers break the `ai.py` ↔ `openai_llm.py` circular dependency

## Planning Document
Implements plan/completed/43-ai-facade-migration.md

## Changes
- `common/ai.py` — rewritten: `AI` is now a facade; all OpenAI logic removed from instance methods
- `common/configuration.py` — three new provider config keys added to defaults and `load_config()`
- `sandvoice.py` — three `AI(self.config)` calls replaced with `AI.from_config(self.config)`
- `tests/test_ai.py` — rewritten to use mock providers; tests `from_config`, provider factories, delegation, and history management
- `tests/test_tts_chunking.py` — `TestTextToSpeechChunking` replaced with delegation-contract tests; detailed chunking tests live in `test_openai_providers.py`

## Testing
- [x] All tests pass: 924 passing
- [x] Coverage >80%
- [x] Tested on Mac M1
- [ ] Tested on Pi 3B

## Configuration Changes
```yaml
llm_provider: openai   # LLM provider (only "openai" supported)
tts_provider: openai   # TTS provider (only "openai" supported)
stt_provider: openai   # STT provider (only "openai" supported)
```

### DIFF
--- a/common/ai.py
+++ b/common/ai.py
@@ -227,15 +227,15 @@ class AI:
         return cls(llm, tts, stt, config)
 
     def generate_response(self, user_input, extra_info=None, model=None):
-        # Pass a snapshot of current history (prior turns) to the provider so that
-        # subsequent mutations to conversation_history don't affect the recorded call.
-        # The provider appends user_input as the final message via _build_messages.
-        result = self._llm.generate_response(
-            user_input, list(self.conversation_history), extra_info=extra_info, model=model
-        )
+        # Append user turn first so history is consistent even if the call fails.
+        # Pass history[:-1] to the provider — provider appends user_input via
+        # _build_messages, avoiding a duplicate if history already ends with this turn.
         user_message = "User: " + user_input
         if not self.conversation_history or self.conversation_history[-1] != user_message:
             self.conversation_history.append(user_message)
+        result = self._llm.generate_response(
+            user_input, self.conversation_history[:-1], extra_info=extra_info, model=model
+        )
         self.conversation_history.append(f"{self.config.botname}: " + result.content)
         return result
 
--- a/common/ai.py
+++ b/common/ai.py
@@ -188,9 +188,10 @@ def _validate_provider_names(config):
 
     Called by from_config before the API-key check so a misconfigured
     provider gives an actionable error rather than a misleading missing-key
-    message.  _build_*_provider functions also validate as a safety net for
-    direct callers; both reference _SUPPORTED_PROVIDERS as the single source
-    of truth.
+    message.  _SUPPORTED_PROVIDERS is the canonical set; _build_*_provider
+    functions also raise ValueError for unknown names as a defensive safety
+    net for direct callers, but they do not reference _SUPPORTED_PROVIDERS
+    directly — update both when adding a new provider.
     """
     for field in ("llm_provider", "tts_provider", "stt_provider"):
         name = getattr(config, field, "openai")
--- a/common/ai.py
+++ b/common/ai.py
@@ -209,7 +209,7 @@ _SUPPORTED_PROVIDERS = {"openai"}
 
 
 class AI:
-    def __init__(self, llm, tts, stt, config, openai_client=None):
+    def __init__(self, llm, tts, stt, config, *, openai_client=None):
         self.config = config
         self._llm = llm
         self._tts = tts
--- a/common/ai.py
+++ b/common/ai.py
@@ -1,8 +1,6 @@
 from openai import OpenAI
-from jinja2 import Template
-import datetime, json, yaml, warnings, os, logging, re, uuid, threading
-from types import SimpleNamespace
-from common.error_handling import retry_with_backoff, setup_error_logging, handle_api_error, handle_file_error
+import os, logging, re
+from common.error_handling import setup_error_logging
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


spideyz0r:

### 1. `generate_response` does not guard against duplicate user turn (line ~230)
- ✅ Mirrored the `stream_response_deltas` pattern: append the user turn to `conversation_history` first, then pass `history[:-1]` to the provider. This prevents a duplicate user message when the caller has already appended the turn (e.g. on retry). The guard `if not self.conversation_history or self.conversation_history[-1] != user_message` skips the append if the turn is already the last entry.

**Test Results:**
- Total tests: 49 (test_ai.py + test_tts_chunking.py)
- All passing ✅

spideyz0r:

### 1. `openai_client` property reaches into provider internals and silently returns `None` (line 219)
- ✅ `AI.__init__` now accepts an optional `openai_client=None` parameter stored as `self._openai_client`. `AI.from_config` passes the constructed client explicitly. The property raises `AttributeError` with a clear message instead of returning `None` when the client is unavailable, decoupling `AI` from provider internals entirely.

### 2. Provider names validated after API key check in `from_config` (line 228)
- ✅ Added upfront provider name validation against `_SUPPORTED_PROVIDERS = {"openai"}` before the `OPENAI_API_KEY` check. A misconfigured `llm_provider`/`tts_provider`/`stt_provider` now raises `ValueError("Unknown <field>: ...")` immediately, before any key check. The set is a module-level constant so it stays in sync with `_build_*_provider` helpers.

### New tests added
- `test_unknown_provider_raises_before_api_key_check` — verifies provider validation fires first even without `OPENAI_API_KEY` set
- `test_openai_client_accessible_after_from_config` — verifies `ai.openai_client` returns the constructed `OpenAI` instance
- `test_openai_client_raises_when_not_set` — verifies `AttributeError` with actionable message when `AI` is constructed directly without a client

**Test Results:**
- Total tests: 927
- All passing ✅

spideyz0r:

### 1. `openai_client` parameter leaks into public positional interface of `AI.__init__` (line 218)
- ✅ Made `openai_client` keyword-only via a `*` separator: `def __init__(self, llm, tts, stt, config, *, openai_client=None)`. The public positional interface remains `(llm, tts, stt, config)` as documented. `from_config` passes it by keyword. No consumer can accidentally rely on a fifth positional argument.

**Test Results:**
- Total tests: 927
- All passing ✅

spideyz0r:

### 1. Unused top-level imports in `tests/test_ai.py` (line 9)
- ✅ Removed the local `from common.providers import ...` re-imports inside each `TestProviderFactories` test method. The module-level imports (`OpenAILLMProvider`, `OpenAITTSProvider`, `OpenAISTTProvider`) are now used directly in the assertions — no dead imports.

### 2. Duplicated provider validation creates sync hazard (line 209)
- ✅ Extracted `_validate_provider_names(config)` as a shared helper (placed alongside `_SUPPORTED_PROVIDERS`). `from_config` now calls `_validate_provider_names` for the early ordering guarantee (providers validated before API key check). `_build_*_provider` functions retain their own `if/raise` as a safety net for direct callers — both reference `_SUPPORTED_PROVIDERS` as the single source of truth.

### 3. Unused `tts` variable in `test_tts_chunking.py` (line 83)
- ✅ Added `tts.text_to_speech.assert_called_once_with(...)` assertions to `test_text_to_speech_multi_chunk` and `test_text_to_speech_returns_empty_on_failure`, making the delegation contract explicit.

**Test Results:**
- Total tests: 927
- All passing ✅

spideyz0r:

### 1. `_validate_provider_names` docstring inaccurately claims `_build_*_provider` references `_SUPPORTED_PROVIDERS` (line 193)
- ✅ Updated the docstring to accurately describe reality: `_SUPPORTED_PROVIDERS` is the canonical set used by `_validate_provider_names`; `_build_*_provider` functions raise `ValueError` for unknown names as a defensive safety net for direct callers but do not reference `_SUPPORTED_PROVIDERS` directly. Added a note that both must be updated when adding a new provider.

**Test Results:**
- Total tests: 927
- All passing ✅

spideyz0r:

### 1. Provider validation duplicated across `_validate_provider_names` and `_build_*_provider` (line 210)
- ✅ Added normalization to `_validate_provider_names`: each value is now stripped, lowercased, and `None`/empty treated as `"openai"` before checking against `_SUPPORTED_PROVIDERS`.
- ✅ Converted `_build_*_provider` helpers from `if provider == "openai": ... raise` chains to explicit dispatch dicts (`_dispatch = {"openai": OpenAILLMProvider}`). The dict keys make the supported names self-documenting and directly correspond to `_SUPPORTED_PROVIDERS`. Adding a new provider means updating `_SUPPORTED_PROVIDERS` + each dispatch dict — both in the same file, clearly documented in the docstring.

### 2. `TestTextToSpeechChunking` class name misaligns with test intent (line 58)
- ✅ Renamed `TestTextToSpeechChunking` → `TestTextToSpeechDelegation` to reflect that these tests verify the delegation contract, not chunking behaviour (which lives in `test_openai_providers.py`).

**Test Results:**
- Total tests: 927
- All passing ✅

spideyz0r:

### 1. `_validate_provider_names` whitespace-only value normalizes to `''` not `"openai"` (line 204)
- ✅ Changed `if raw` truthiness guard to `if raw is not None`, then added `or "openai"` fallback after stripping. Whitespace-only strings like `"   "` now strip to `""`, which falls through to the default — consistent with `None` and `""`.
- Before: `str(raw).strip().lower() if raw else "openai"` — `"   "` is truthy, evaluates to `""`, raises `Unknown llm_provider: ''`
- After: `(str(raw).strip().lower() if raw is not None else "") or "openai"`

### 2. `_build_tts_provider` non-string truthy value raises `AttributeError` (line 226)
- ✅ Applied the same normalization pattern as `_validate_provider_names`: extract `raw`, then `(str(raw).strip().lower() if raw is not None else "") or "openai"`. Non-string truthy values now coerce to `str` and produce a clear `ValueError("Unknown tts_provider: ...")` instead of `AttributeError`.

### 3. `_build_stt_provider` same normalization issue (line 236)
- ✅ Same fix as #2.

### Bonus fix (not flagged, same root cause)
- ✅ `_build_llm_provider` had the identical pattern at line 212 — fixed proactively for consistency.

### New test
- ✅ Added `test_whitespace_only_provider_defaults_to_openai`: verifies that `llm_provider="   "`, `tts_provider="   "`, `stt_provider="   "` all normalize to `"openai"` without raising.

**Test Results:**
- Total tests: 928
- All passing ✅

PR #128, opened 2026-04-13
Add plan 47: extract system prompt to common/prompt.py

spideyz0r:
## Summary

Adds Plan 47 to the backlog: extract the SandVoice system prompt out of `OpenAILLMProvider` and into a shared `common/prompt.py` utility function.

## Motivation

`OpenAILLMProvider._build_system_role()` contains the core SandVoice persona — botname, language, timezone, TTS formatting rules, verbosity — none of which is OpenAI-specific. Any future provider (Gemini, Anthropic, etc.) would need the same prompt, forcing duplication.

Extracting it to `build_system_role(config, extra_info=None)` in `common/prompt.py` gives every provider a single import with no inheritance required.

## No code changes

Plan document only. Implementation follows in a separate PR.

### DIFF
--- a/plan/INDEX.md
+++ b/plan/INDEX.md
@@ -214,6 +214,10 @@ plan/
 **Document**: [backlog/46-remove-openai-client-escape-hatch.md](./backlog/46-remove-openai-client-escape-hatch.md)
 **Description**: Migrate `voice_filler.py` to `ai.one_shot()` and `realtime_websearch/plugin.py` to `ai.web_search()`, then remove the `openai_client` property and `_openai_client` attribute from `AI` entirely. After this plan, no plugin or instance method touches the OpenAI SDK directly. Requires Plans 44 and 45.
 
+### Priority 47: Extract System Prompt to common/prompt.py
+**Document**: [backlog/47-system-prompt-extraction.md](./backlog/47-system-prompt-extraction.md)
+**Description**: Move `OpenAILLMProvider._build_system_role()` logic into a standalone `build_system_role(config, extra_info=None)` function in `common/prompt.py`. The SandVoice bot identity, language, timezone/location context, verbosity instructions, and the formatting constraint not to reply as a chat are application-level concerns — not OpenAI-specific. Any future provider imports and calls the shared function directly.
+
 ### Future Enhancements
 **Document**: [backlog/FUTURE.md](./backlog/FUTURE.md)
 **Description**: Long-term feature ideas including API Cost Management, Conversation History Management, Code Deduplication, Timers & Reminders, Music Control, Smart Home Integration, Calendar Integration, Todo List Management, Multi-User Support, and Conversation Memory.
--- a/plan/backlog/47-system-prompt-extraction.md
+++ b/plan/backlog/47-system-prompt-extraction.md
@@ -1,0 +1,130 @@
+# Plan 47: Extract System Prompt to common/prompt.py
+
+## Problem
+
+`OpenAILLMProvider._build_system_role()` contains the core SandVoice persona and
+conversation instructions:
+
+```python
+system_role = f"""
+    Your name is {self.config.botname}.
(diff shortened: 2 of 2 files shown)


spideyz0r:

### 1. Problem snippet includes "Never use symbols" not yet in main (line 24)
- ✅ Removed the hardcoded snippet entirely; replaced with a prose description of what the method contains (bot identity, language, timezone/location, verbosity, chat-format constraint). Added a note that the implementer must check `openai_llm.py` at implementation time and include any instructions added by later PRs (e.g. #127).

### 2. Proposed `build_system_role()` sample includes "Never use symbols" (line 75)
- ✅ Removed that line from the sample. The sample now matches the current `_build_system_role()` on main exactly.

### 3. `text_summary` incorrectly listed as a call site (line 100)
- ✅ Removed `text_summary` from the call-site list. It builds its own separate system role and does not call `_build_system_role`.

### 4. INDEX description mentions "TTS formatting rules" not in current method (INDEX line 207)
- ✅ Replaced with an accurate description of what the current method actually contains: bot identity, language, timezone/location context, verbosity instructions, and the formatting constraint not to reply as a chat.

**No code changes — plan documents only.**

spideyz0r:

### 1. "verbatim" conflicts with the `self.config` → `config` substitution (line 28)
- ✅ Replaced "copy the body verbatim" with clearer wording: "copy with only the minimal `self.config` → `config` substitution needed for the extraction; otherwise preserve the logic and prompt text exactly". Also added an explicit callout that the leading/trailing whitespace inside the triple-quoted f-string is part of the prompt and must not be cleaned up during the move.

### 2. Whitespace in f-string not called out; tests could miss it (line 66)
- ✅ Added a note to the Approach section that whitespace must be preserved exactly, and updated the acceptance criteria to require **exact string equality** assertions in `tests/test_prompt.py` (not `assertIn`) so any accidental indentation change is caught.

**No code changes — plan documents only.**

spideyz0r:

### 1. Sample f-string indentation doesn't match `openai_llm.py` (line 74)
- ✅ Removed the illustrative code snippet entirely. Added an explicit note: **"Do not rely on the snippet below for exact indentation or content — the source of truth is `common/providers/openai_llm.py` at implementation time. Copy/paste the f-string directly from that file."** This avoids any risk of the implementer following a slightly-off plan snippet rather than the actual source.

### 2. Exact equality tests won't work with `datetime.datetime.now()` (line 45)
- ✅ Added to acceptance criteria: "patch `datetime.datetime.now` to a fixed value so assertions are deterministic".

### 3. "Pure function" claim is inaccurate due to `datetime.now()` (line 126)
- ✅ Replaced "pure function" with accurate description: "reads config values and returns a string; calls `datetime.datetime.now()` internally so it is time-dependent (not strictly pure), but has no other side effects. Tests must patch `datetime.datetime.now` to a fixed value."

**No code changes — plan documents only.**

spideyz0r:

### 1. Problem statement omits TTS formatting instruction (line 12)
- ✅ Updated the problem statement to accurately describe all current contents of `_build_system_role()`: bot identity, language, timezone/location, current date/time, verbosity, chat-format constraint, and the TTS symbol-spelling instruction.
- Note: the TTS formatting instruction ("Never use symbols...") was added by PR #127, which merged before this review — so it is now present on main. The previous description predated that merge.

**No code changes — plan documents only.**

spideyz0r:

### 1. Problem statement mentions TTS symbol-spelling instruction not in base branch (line 9)
- ✅ Removed the TTS instruction mention from the problem statement. That instruction was added by PR #127, which merged after this PR was opened — so Copilot correctly flags it as not present in the base snapshot it reviews against.
- The existing note in the Approach section already covers this: "check for any instructions added by later PRs (e.g. PR #127) and include them" — so implementers will still pick it up at implementation time.

**No code changes — plan documents only.**

PR #137, opened 2026-04-19
Implement Plan 51: greeting_extra config key

spideyz0r:
## Summary
- Add optional `greeting_extra` config key that appends user-defined instructions to the greeting plugin's generation prompt (e.g. "end the greeting with a short proverb")
- Appended to `extra_system` after `.rstrip()` so the injected text always starts on a clean line
- Blank/whitespace-only or non-string values normalised to `None` with a warning logged

## Planning Document
Implements plan/backlog/51-greeting-extra-instructions.md

## Changes
- `common/configuration.py`: add `greeting_extra` to defaults; parse/validate in `load_config()`
- `plugins/greeting/plugin.py`: inject `greeting_extra` into `extra_system` before LLM call
- `README.md`: add commented-out example in config section
- `tests/test_greeting_plugin.py`: `TestGreetingExtra` (3 cases)
- `tests/test_configuration.py`: `TestGreetingExtraConfig` (6 cases)

## Testing
- [x] All tests pass (978 passed)
- [x] Coverage 88%
- [x] Tested on Mac M1

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,9 @@ voice_filler_phrases:                # set to [] to disable
   - "Let me check that."
   - "Sure, one moment."
 
+# greeting_extra: |             # optional: append custom instructions to the greeting prompt
+#   End the greeting with a short, relevant proverb.
+
 # Scheduler (see Scheduled Tasks section)
 scheduler_enabled: disabled
 scheduler_poll_interval: 30
--- a/common/configuration.py
+++ b/common/configuration.py
@@ -368,10 +368,16 @@ class Config:
         raw_ge = self.get("greeting_extra")
         if raw_ge is None:
             self.greeting_extra = None
-        elif not isinstance(raw_ge, str) or not raw_ge.strip():
+        elif not isinstance(raw_ge, str):
             logger.warning(
                 "greeting_extra must be a non-empty string; ignoring value of type %s",
                 type(raw_ge).__name__,
+            )
+            self.greeting_extra = None
+        elif not raw_ge.strip():
+            logger.warning(
+                "greeting_extra is blank or whitespace-only (length=%d); ignoring",
+                len(raw_ge),
             )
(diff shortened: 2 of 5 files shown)


spideyz0r:

### 1. `plugins/greeting/plugin.py` line 93 — `greeting_extra.strip()` called three times
- ✅ Stored `greeting_extra.strip()` once as `greeting_extra_stripped`; used in concatenation and `len()` call

### 2. `common/configuration.py` line 352 — Warning misleading for blank strings
- ✅ Split into two branches: non-string logs `"ignoring value of type %s"`, blank/whitespace logs `"is blank or whitespace-only (length=%d)"`
- ✅ Added `test_blank_string_logs_warning` to verify the new distinct message

### 3. `tests/test_configuration.py` line 1249 — Duplicate temp-HOME helpers
- ✅ Changed `TestGreetingExtraConfig(unittest.TestCase)` → `TestGreetingExtraConfig(_TempHomeBase)` and removed duplicated `setUp`/`tearDown`/`write_config` methods

### 4. `tests/test_configuration.py` line 1243 — `tearDown` uses `if self.original_home:` instead of `is not None`
- ✅ Resolved by item 3: `TestGreetingExtraConfig` no longer has its own `tearDown`; it inherits from `_TempHomeBase`

### 5. `tests/test_greeting_plugin.py` line 389 — Fragile `assertNotIn("proverb", ...)`
- ✅ Replaced with `assertNotIn("End the greeting with a short proverb.", ...)` — the specific `greeting_extra` string from the positive test, not an incidental word from the base prompt

### 6. `tests/test_greeting_plugin.py` line 412 — Weak `endswith("   \n")` assertion
- ✅ Replaced with a `_capture()` helper that runs the plugin twice (once with `None`, once with `"   "`), then asserts the captured prompts are identical

**Test Results:**
- 979 passed ✅

PR #139, opened 2026-04-20
Implement Plan 37: context-aware routing via conversation history

spideyz0r:
## Summary
- Passes the last N conversation turns to `define_route` so the routing LLM can resolve follow-up utterances and corrections in context
- Fixes misrouting of clarifications like "please answer in Portuguese" after a previous query — without history the router has no context and picks a wrong plugin

## Planning Document
Implements `plan/backlog/37-context-aware-routing.md`

## Changes
- `common/providers/base.py`: add `history=None` to `LLMProvider.define_route` ABC
- `common/providers/openai_llm.py`: prepend history entries as user messages in `_call_define_route`; log history depth at INFO
- `common/ai.py`: thread `history` parameter through `AI.define_route`
- `common/configuration.py`: add `route_history_depth` config key (default 4, 0 disables)
- `sandvoice.py` + `common/wake_word.py`: pass `conversation_history[-depth:]` to `define_route`
- `tests/test_openai_providers.py`: 3 new tests (no history, history prepended, empty history)
- `tests/test_configuration.py`: 5 new tests for `route_history_depth` (default, custom, zero, negative, invalid)
- `tests/test_ai.py` + `tests/test_wake_word.py`: updated call-site assertions

## Stateless callers unaffected
`greeting.plugin` and scheduler both call `define_route` without `history` (default `None`) — their routing is already correct because they always send self-contained queries.

## Testing
- [x] All tests pass (1002 passed)
- [x] Coverage >80%
- [x] Tested on Mac M1

### DIFF
--- a/common/ai.py
+++ b/common/ai.py
@@ -319,8 +319,8 @@ class AI:
     def transcribe_and_translate(self, model=None, audio_file_path=None):
         return self._stt.transcribe(audio_file_path=audio_file_path, model=model)
 
-    def define_route(self, user_input, model=None, extra_routes=None):
-        return self._llm.define_route(user_input, model=model, extra_routes=extra_routes)
+    def define_route(self, user_input, model=None, extra_routes=None, history=None):
+        return self._llm.define_route(user_input, model=model, extra_routes=extra_routes, history=history)
 
     def text_to_speech(self, text, model=None, voice=None):
         return self._tts.text_to_speech(text, model=model, voice=voice)
--- a/common/configuration.py
+++ b/common/configuration.py
@@ -388,10 +388,14 @@ class Config:
         else:
             self.greeting_extra = raw_ge.strip()
 
-        try:
-            self.route_history_depth = max(0, int(self.get("route_history_depth")))
-        except (ValueError, TypeError):
-            logger.warning("route_history_depth must be a non-negative integer; using default 4")
+        _raw_depth = _parse_exact_int(self.get("route_history_depth"))
+        if isinstance(_raw_depth, int) and not isinstance(_raw_depth, bool):
+            self.route_history_depth = max(0, _raw_depth)
+        else:
+            logger.warning(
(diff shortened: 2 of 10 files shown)


spideyz0r:

### 1. `route_history_depth` accepts booleans and non-integer floats (configuration.py line 394) — Fixed
- ✅ Switched from `int(self.get(...))` to `_parse_exact_int()` (already used elsewhere in Config) which rejects bools and non-exact floats
- ✅ Warning now logs the default value from `self.defaults["route_history_depth"]` instead of the hardcoded `4`
- ✅ Added `test_bool_rejected_falls_back_to_default` and `test_float_rejected_falls_back_to_default` tests

### 2. Duplicated depth parsing / empty-slice not normalized to None (sandvoice.py line 649) — Fixed
- ✅ Consolidated into a single `try/except TypeError` block covering both `max(0, ...)` and the slice
- ✅ Empty slice normalized to `None` with `or None` so `define_route` never receives `history=[]`

### 3. Same consolidation for wake_word.py (line 558) — Fixed
- ✅ Same pattern applied; single `try/except TypeError` block

### 4. Current `user_input` sent unprefixed while history entries carry "User: ..." prefixes (openai_llm.py line 150) — Fixed
- ✅ When `history` is present, current input is sent as `"User: {user_input}"` to match the format of history entries
- ✅ When `history` is `None` (stateless callers), input is sent as-is (no change to existing behaviour)
- ✅ Updated `test_history_prepended_before_current_input` to assert `"User: what is the weather?"`

**Test Results:**
- Total tests: 1004
- All passing ✅

spideyz0r:

### 1. Inconsistent user_input format with/without history (openai_llm.py) — Fixed
- ✅ Removed the `if history / else` split; `user_input` is now always sent as `"User: {user_input}"` regardless of whether history is present
- ✅ Updated `test_no_history_sends_system_and_user` to assert the prefixed format

**Test Results:**
- Total tests: 1004
- All passing ✅

PR #141, opened 2026-04-21
Add Plan 54 and 55: conversation history SQLite + Telegram channel

spideyz0r:
## Summary
- **Plan 54**: Persist conversation turns to SQLite so history survives restarts and can be shared across channels. New `ConversationHistory` class, `history_enabled` / `history_max_entries` config keys.
- **Plan 55**: Telegram channel as a background thread — text in/out, user ID whitelist, shared `ConversationHistory` with wake-word session. Phase 1 text-only; Phase 2 adds voice. New `telegram_enabled`, `telegram_bot_token`, `telegram_allowed_user_ids` config keys.

## Planning Documents
- `plan/backlog/54-conversation-history-sqlite.md`
- `plan/backlog/55-telegram-channel.md`
- `plan/INDEX.md`: adds Priority 54 and 55 entries

## Design decisions
- Plan 54 is a hard dependency for Plan 55 — shared history requires SQLite persistence
- No turn lock: occasional context interleave from parallel channels is acceptable
- Telegram runs as a daemon thread (same `TaskScheduler` pattern), not a separate process
- Phase 1 is text-only to keep scope small and validate the architecture

## Testing
N/A — docs-only PR

### DIFF
--- a/plan/INDEX.md
+++ b/plan/INDEX.md
@@ -247,6 +247,14 @@ plan/
 **Document**: [backlog/55-vad-pre-speech-grace-period.md](./backlog/55-vad-pre-speech-grace-period.md)
 **Description**: Only start the VAD silence countdown after at least one speech frame has been detected. Prevents the 1.5 s silence window from cutting off the recording before the user has had time to start speaking after the wake word beep.
 
+### Priority 56: Conversation History Persistence (SQLite)
+**Document**: [backlog/54-conversation-history-sqlite.md](./backlog/54-conversation-history-sqlite.md)
+**Description**: Persist each conversation turn to a `conversation_history` SQLite table in the existing DB file. On startup, seed the in-memory list from the last `history_max_entries` rows so sessions resume naturally. Required foundation for multi-channel history sharing (Plan 57).
+
+### Priority 57: Telegram Channel
+**Document**: [backlog/55-telegram-channel.md](./backlog/55-telegram-channel.md)
+**Description**: Add an always-on Telegram channel as a background thread within the SandVoice process. Text messages from whitelisted user IDs are routed through the same AI, plugins, and conversation history as wake-word mode. Adds `telegram_enabled`, `telegram_bot_token`, and `telegram_allowed_user_ids` config keys. Phase 1 is text only; Phase 2 adds voice message support.
+
 ### Future Enhancements
 **Document**: [backlog/FUTURE.md](./backlog/FUTURE.md)
 **Description**: Long-term feature ideas: alternative providers (Anthropic, Ollama, Piper), local offline mode, conversation history truncation, user-triggered timers, conversation memory, export, plugin hot-reload, API cost tracking, music control, smart home, calendar, todo lists, multi-user support.
--- a/plan/backlog/54-conversation-history-sqlite.md
+++ b/plan/backlog/54-conversation-history-sqlite.md
@@ -27,8 +27,9 @@ table, giving them a shared, durable conversation history.
   channel runs as a background thread within the same process, not a separate process,
   so no multi-process SQLite guarantees are required.
 - History search, export, or summarisation — future work.
-- Pruning old entries beyond the startup load window — simple mtime/rowid pruning on
-  startup is sufficient.
+- Pruning old entries — rows accumulate indefinitely; a future plan can add a cleanup
+  job. The startup load window (`history_max_entries`) limits memory use without
+  touching the DB.
 
 ## Design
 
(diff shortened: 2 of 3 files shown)


spideyz0r:

### 1. `config.db_path` doesn't exist; AI construction is wrong (Plan 54 line 88) ✅
- ✅ Changed `config.db_path` → `config.scheduler_db_path` (the shared SQLite file used by `VoiceCache` and `SchedulerDB`)
- ✅ Updated AI construction comment to note that the OpenAI client is internal — `AI(config, history=history)`

### 2. History prefix uses `"Assistant:"` but codebase uses `config.botname` (Plan 54 line 61) ✅
- ✅ Updated `load_recent` description to use `"<botname>: ..."` with a note that botname comes from `config.botname` (e.g. `"Sandbot: ..."`)

### 3. Out-of-scope wording implies separate processes (Plan 54 line 27) ✅
- ✅ Clarified: Plan 55's Telegram channel runs as a background thread within the same process — no multi-process SQLite guarantees needed

### 4. Message flow uses `route in plugins` instead of `route["route"]` (Plan 55 line 56) ✅
- ✅ Fixed flow to: `route_name = route["route"]`, then `plugins[route_name](user_input, route, ctx)` — matches `SandVoice.route_message` dispatch pattern

### 5. `run_polling()` installs signal handlers that fail in non-main thread (Plan 55 line 48) ✅
- ✅ Added explicit note: must pass `stop_signals=()` to `run_polling()`, or use lower-level async lifecycle methods (`initialize()`/`start()`/`updater.start_polling()`)

### 6. Thread-safety: SQLite persistence doesn't make in-memory list thread-safe (Plan 55 line 65) ✅
- ✅ Clarified design: Telegram thread does NOT share `AI.conversation_history` in memory — it reads context fresh from DB per turn and writes via `ConversationHistory.append()`. SQLite WAL handles concurrent writes. No turn lock needed for this architecture.

**Test Results:**
- Total tests: 1004
- All passing ✅

spideyz0r:

### 1. Startup uses `AI(config, history=history)` but real constructor takes provider objects (Plan 54 line 90) ✅
- ✅ Fixed startup snippet to use `AI.from_config(config)` (the real factory), then seed `ai.conversation_history` directly from `history.load_recent()` after construction. Added note explaining that `AI.__init__` takes `(llm, tts, stt, config)`, so callers must go through `from_config`. Also noted the alternative of extending `from_config` to accept `history` as an optional param.

### 2. `TelegramChannel` constructor missing `history` parameter (Plan 55 line 41) ✅
- ✅ Added `history` to constructor signature: `__init__(self, config, ai, plugins, cache, history)`
- ✅ Updated startup snippet to pass `history` when constructing `TelegramChannel`

### 3. Message flow calls `ai.define_route(user_input)` with no history (Plan 55 line 64) ✅
- ✅ Added `recent_history = history.load_recent(config.route_history_depth)` before routing
- ✅ Updated to `ai.define_route(user_input, history=recent_history)`

### 4. `ai.generate_response(user_input)` returns a result object, not a string (Plan 55 line 64) ✅
- ✅ Fixed to `ai.generate_response(user_input).content` — consistent with how `sandvoice.py` uses the return value

**Test Results:**
- Total tests: 1004
- All passing ✅

spideyz0r:

### 1. Startup only seeds history but doesn't hook future appends into AI (Plan 54 line 92) ✅
- ✅ Redesigned: `AI.from_config` extended to accept optional `history` parameter. `AI.__init__` stores it as `self._history` and calls `history.append()` inside `append_to_history()` on every turn. Seeding **and** ongoing persistence both live inside `AI` — no external caller can double-write.

### 2. Pruning out-of-scope bullet is internally contradictory (Plan 54 line 31) ✅
- ✅ Clarified: pruning is fully out of scope. Rows accumulate indefinitely; a future plan can add cleanup. `history_max_entries` limits memory use without touching the DB.

### 3. Plan 55 says "same AI" but AI.conversation_history would be shared and mutated concurrently (Plan 55 line 26) ✅
- ✅ Redesigned: `TelegramChannel` creates its **own** `AI` instance inside the background thread via `AI.from_config(config, history=history)`. Both threads share `ConversationHistory` (SQLite + `threading.Lock`) but each have their own in-memory list — no concurrent mutation.

### 4. Double-write risk: message flow calls both `ai.generate_response()` and `history.append()` (Plan 55 line 68) ✅
- ✅ Removed manual `history.append()` calls from message flow. `AI.append_to_history()` is the single source of truth — persistence happens there, same as the main thread.

### 5. "SQLite WAL handles concurrent writes" misleading for in-process threading (Plan 55 line 76) ✅
- ✅ Clarified: `ConversationHistory` uses a `threading.Lock` (same pattern as `VoiceCache`). WAL alone is not sufficient for in-process thread safety.

### 6. No behaviour defined when `telegram_enabled` but `history_enabled: disabled` (Plan 55 line 94) ✅
- ✅ Added startup guard: if `telegram_enabled` without `history_enabled`, log a warning and skip starting the Telegram channel.

**Test Results:**
- Total tests: 1004
- All passing ✅

PR #148, opened 2026-05-02
Replace Porcupine with openWakeWord (Plan 52)

spideyz0r:
## Summary
- Replaces pvporcupine with openWakeWord (MIT-licensed, ONNX-based, no API key required)
- New `common/openwakeword_detector.py`: thin wrapper with Porcupine-compatible interface — `process(pcm)`, `reset()`, `delete()`, `frame_length`/`sample_rate` properties
- Handles resampling (48 kHz Pi → 16 kHz model) using precomputed gcd-based ratios via `scipy.signal.resample_poly`
- Adds `_find_hw_input_device()` in wake_word.py for correct Linux hw: device selection
- Fixes VAD grace period in vad_recorder.py: silence timer only starts after speech is detected
- Adds SDL ALSA output device detection in audio.py for correct Linux audio routing

## Planning Document
Implements plan/backlog/52-replace-porcupine-with-openwakeword.md

## Changes
- `common/openwakeword_detector.py` — new detector class
- `common/wake_word.py` — switch from pvporcupine to OpenWakeWordDetector; add hw input device selection; beep after stream close
- `common/barge_in.py` — same swap; update signature to `model_name`/`threshold`
- `common/vad_recorder.py` — VAD grace period (Plan 55 feature)
- `common/audio.py` — SDL ALSA device detection on Linux
- `common/configuration.py` — drop porcupine_access_key/keyword_paths, add openwakeword_model
- `requirements.txt` — remove pvporcupine, add openwakeword>=0.6.0, onnxruntime==1.20.0, scipy
- `sandvoice.py` — update import error message
- `README.md` — remove Porcupine config keys, document new config
- `tests/test_openwakeword_detector.py` — new tests (>80% coverage)
- `tests/test_barge_in.py`, `tests/test_wake_word.py`, `tests/test_configuration.py` — updated for new interfaces

## Testing
- [x] All 1056 tests pass
- [x] Coverage >80% for new code
- [x] Tested on Mac M1
- [ ] Tested on Pi 3B

## Configuration Changes
- `openwakeword_model: hey_jarvis` — model name or absolute path to `.onnx` file
- `wake_phrase: hey jarvis` — updated default
- `wake_word_sensitivity: 0.35` — updated default (openWakeWord range)
- Removed: `porcupine_access_key`, `porcupine_keyword_paths`

## Pi 3B Install Note
`pip install openwakeword>=0.6.0 --no-deps && pip install scipy`
(tflite-runtime has no aarch64 wheel for Python 3.13; onnxruntime pinned to 1.20.0 — 1.21+ crashes with STL vector assertion)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,6 +31,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install "setuptools<81"  # Required for pkg_resources used by webrtcvad (removed in v81+)
+          pip install openwakeword>=0.6.0 --no-deps  # tflite-runtime has no Linux wheel for Python 3.10-3.12
           pip install -r requirements.txt
 
       - name: Run tests with coverage
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,15 +31,9 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install "setuptools<81"  # Required for pkg_resources used by webrtcvad (removed in v81+)
-          # openwakeword declares tflite-runtime as a Linux dep, but has no wheel for
+          # openwakeword declares tflite-runtime as a Linux dep but no wheel exists for
           # Python 3.10-3.12. We use ONNX inference only, so install a stub to satisfy pip.
-          python -c "
-import tempfile, os, subprocess, sys
-td = tempfile.mkdtemp()
-with open(os.path.join(td, 'setup.py'), 'w') as f:
-    f.write(\"from setuptools import setup; setup(name='tflite-runtime', version='2.14.0', packages=[])\")
-subprocess.run([sys.executable, '-m', 'pip', 'install', td], check=True)
-"
+          TD=$(mktemp -d) && echo "from setuptools import setup; setup(name='tflite-runtime', version='2.14.0', packages=[])" > $TD/setup.py && pip install $TD
           pip install -r requirements.txt
 
       - name: Run tests with coverage
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,7 +31,15 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install "setuptools<81"  # Required for pkg_resources used by webrtcvad (removed in v81+)
-          pip install openwakeword>=0.6.0 --no-deps  # tflite-runtime has no Linux wheel for Python 3.10-3.12
+          # openwakeword declares tflite-runtime as a Linux dep, but has no wheel for
+          # Python 3.10-3.12. We use ONNX inference only, so install a stub to satisfy pip.
+          python -c "
+import tempfile, os, subprocess, sys
+td = tempfile.mkdtemp()
+with open(os.path.join(td, 'setup.py'), 'w') as f:
+    f.write(\"from setuptools import setup; setup(name='tflite-runtime', version='2.14.0', packages=[])\")
+subprocess.run([sys.executable, '-m', 'pip', 'install', td], check=True)
+"
           pip install -r requirements.txt
 
       - name: Run tests with coverage
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ rss_news_max_items: "5"
 # Wake word (--wake-word mode)
 wake_word_enabled: enabled   # global toggle; --wake-word flag is still required to start the mode
 wake_phrase: hey jarvis
-wake_word_sensitivity: 0.35  # 0.0-1.0; higher = more sensitive
+wake_word_sensitivity: 0.35  # 0.0-1.0; higher = stricter (fewer false positives), lower = more sensitive
 openwakeword_model: hey_jarvis  # built-in model name or path to a .onnx file
 
 # Voice activity detection (required in --wake-word mode)
--- a/README.md
+++ b/README.md
@@ -51,10 +51,9 @@ bot_voice: enabled
 stream_responses: enabled
 stream_tts: enabled
 vad_enabled: enabled
-porcupine_access_key: "YOUR_KEY_HERE"   # free key at https://console.picovoice.ai/
 ```
 
-The wake phrase defaults to `hey sandvoice`. Change it with `wake_phrase` in config (must be a built-in Porcupine keyword, or provide a custom `.ppn` model via `porcupine_keyword_paths`).
+The wake phrase defaults to `hey jarvis`. Change it with `wake_phrase` and `openwakeword_model` in config. The model name must be a built-in openWakeWord model (e.g. `hey_jarvis`, `alexa`) or an absolute path to a custom `.onnx` file.
 
 ## API keys
 
@@ -62,7 +61,6 @@ The wake phrase defaults to `hey sandvoice`. Change it with `wake_phrase` in con
 |---|---|---|
 | `OPENAI_API_KEY` | Always | All AI features |
 | `OPENWEATHERMAP_API_KEY` | Weather plugin only | `weather` plugin |
-| `porcupine_access_key` (in config) | Wake word mode only | Wake word, barge-in |
 
 ## Configuration
 
@@ -131,10 +129,9 @@ rss_news_max_items: "5"
 
 # Wake word (--wake-word mode)
 wake_word_enabled: enabled   # global toggle; --wake-word flag is still required to start the mode
-wake_phrase: hey sandvoice
-wake_word_sensitivity: 0.5   # 0.0-1.0; higher = more sensitive
-porcupine_access_key: ""
-porcupine_keyword_paths: null   # path(s) to custom .ppn model(s), or null
+wake_phrase: hey jarvis
+wake_word_sensitivity: 0.35  # 0.0-1.0; higher = more sensitive
+openwakeword_model: hey_jarvis  # built-in model name or absolute path to a custom .onnx file
 
 # Voice activity detection (required in --wake-word mode)
 vad_enabled: enabled
--- a/README.md
+++ b/README.md
@@ -55,6 +55,13 @@ vad_enabled: enabled
 
 The wake phrase defaults to `hey jarvis`. Change it with `wake_phrase` and `openwakeword_model` in config. The model name must be a built-in openWakeWord model (e.g. `hey_jarvis`, `alexa`) or a path to a `.onnx` file.
 
+**Raspberry Pi / Linux install note:** `tflite-runtime` (a transitive dependency of openWakeWord) has no wheels for Python 3.10–3.12 on Linux. Install a no-op stub before running `pip install -r requirements.txt`:
+
+```bash
+TD=$(mktemp -d) && echo "from setuptools import setup; setup(name='tflite-runtime', version='2.14.0', packages=[])" > $TD/setup.py && pip install $TD
+pip install -r requirements.txt
+```
+
 ## API keys
 
 | Key | Required | Used by |
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ vad_enabled: enabled
 
 The wake phrase defaults to `hey jarvis`. Change it with `wake_phrase` and `openwakeword_model` in config. The model name must be a built-in openWakeWord model (e.g. `hey_jarvis`, `alexa`) or a path to a `.onnx` file.
 
-**Raspberry Pi / Linux install note:** `tflite-runtime` (a transitive dependency of openWakeWord) has no wheels for Python 3.10–3.12 on Linux. Install a no-op stub before running `pip install -r requirements.txt`:
+**Raspberry Pi / Linux install note:** `tflite-runtime` (a transitive dependency of openWakeWord) has no compatible wheels for current Python versions on Linux. Install a no-op stub before running `pip install -r requirements.txt`:
 
 ```bash
 TD=$(mktemp -d) && echo "from setuptools import setup; setup(name='tflite-runtime', version='2.14.0', packages=[])" > $TD/setup.py && pip install $TD
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ stream_tts: enabled
 vad_enabled: enabled
 ```
 
-The wake phrase defaults to `hey jarvis`. Change it with `wake_phrase` and `openwakeword_model` in config. The model name must be a built-in openWakeWord model (e.g. `hey_jarvis`, `alexa`) or an absolute path to a custom `.onnx` file.
+The wake phrase defaults to `hey jarvis`. Change it with `wake_phrase` and `openwakeword_model` in config. The model name must be a built-in openWakeWord model (e.g. `hey_jarvis`, `alexa`) or a path to a `.onnx` file.
 
 ## API keys
 
@@ -131,7 +131,7 @@ rss_news_max_items: "5"
 wake_word_enabled: enabled   # global toggle; --wake-word flag is still required to start the mode
 wake_phrase: hey jarvis
 wake_word_sensitivity: 0.35  # 0.0-1.0; higher = more sensitive
-openwakeword_model: hey_jarvis  # built-in model name or absolute path to a custom .onnx file
+openwakeword_model: hey_jarvis  # built-in model name or path to a .onnx file
 
 # Voice activity detection (required in --wake-word mode)
 vad_enabled: enabled
--- a/common/audio.py
+++ b/common/audio.py
@@ -1,14 +1,43 @@
 #import re, os, pyaudio, wave
 #from pydub import AudioSegment
 import contextlib
-import os, platform, time, threading, pyaudio, wave, lameenc, logging, queue
-from pynput import keyboard
+import os, platform, time, threading, pyaudio, wave, lameenc, logging, queue, re
 from common.error_handling import handle_file_error
 
 logger = logging.getLogger(__name__)
 
 # this is necessary to mute some outputs from pygame
 os.environ['PYGAME_HIDE_SUPPORT_PROMPT'] = "hide"
+
+# On Linux, SDL defaults to the first ALSA device which may not be the USB audio device.
+# Point SDL at the first real hardware output device (name contains "hw:N,M").
+# Prefer a combined in+out device (USB headset) over output-only.
+# Falls back to "default" if no hardware device is found.
+# Users can override by setting SDL_AUDIODRIVER / AUDIODEV in their environment.
+if platform.system().lower() == "linux" and "SDL_AUDIODRIVER" not in os.environ:
+    os.environ["SDL_AUDIODRIVER"] = "alsa"
+    try:
+        _pa = pyaudio.PyAudio()
+        try:
+            _audiodev = None
+            _fallback = None
+            for _i in range(_pa.get_device_count()):
+                _dev = _pa.get_device_info_by_index(_i)
+                _m = re.search(r'hw:(\d+),(\d+)', _dev.get("name", ""))
+                if not _m:
+                    continue
+                _plug = f"plughw:{_m.group(1)},{_m.group(2)}"
+                if _dev.get("maxOutputChannels", 0) > 0 and _dev.get("maxInputChannels", 0) > 0:
+                    _audiodev = _plug
+                    break
+                if _fallback is None and _dev.get("maxOutputChannels", 0) > 0:
+                    _fallback = _plug
+            os.environ["AUDIODEV"] = _audiodev or _fallback or "default"
+        finally:
+            _pa.terminate()
+    except Exception:
+        os.environ["AUDIODEV"] = "default"
+
 import pygame
 
 
@@ -55,6 +84,7 @@ class Audio:
             logger.debug(">>> MIXER STATE [%s]: Error getting state: %s", context, e)
 
     def init_recording(self):
+        from pynput import keyboard
         listener = keyboard.Listener(on_press=self.on_press)
         listener.start()
         self.start_recording()
--- a/common/audio.py
+++ b/common/audio.py
@@ -9,6 +9,30 @@ logger = logging.getLogger(__name__)
 # this is necessary to mute some outputs from pygame
 os.environ['PYGAME_HIDE_SUPPORT_PROMPT'] = "hide"
 
+# Kept at module scope so the ctypes callback is never garbage-collected
+# while ALSA may still invoke it (local variables would be GC'd on return).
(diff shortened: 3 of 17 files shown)


spideyz0r:

### 1. Clip resampled audio to int16 range (`common/openwakeword_detector.py` line 97)
- ✅ Added `np.clip(..., -32768, 32767)` before `.astype(np.int16)` after `resample_poly`. Filter ringing in the polyphase resampler can push values outside the int16 range; direct cast would silently wrap, distorting the signal fed to the model.

### 2. VAD no-speech latency regression (`common/vad_recorder.py` line 133)
- ✅ Added a pre-speech timeout: when `speech_detected` is still False, the loop breaks after `vad_silence_duration` seconds have elapsed since recording started. Also added `not speech_detected` to the early-return check so captured silence is discarded (returns None) rather than sent for transcription. This preserves the grace-period behaviour (silence timer only starts after speech for the post-speech cutoff) while fixing the regression where a missed wake-word would hold the mic open for the full `vad_timeout` (30 s).

### 3. Barge-in stream rate/frame mismatch (`common/barge_in.py` line 256)
- ✅ Fixed: open the PyAudio stream at `detector_instance.device_sample_rate` (the actual capture rate, e.g. 48 kHz on Pi) instead of `detector_instance.sample_rate` (always 16 kHz). `frame_length` is scaled to the device rate, so opening at 16 kHz would read the wrong number of samples per frame.

### 4. Barge-in missing Linux hw: device selection (`common/barge_in.py` line 256)
- ✅ Added `_find_hw_input_device(pa)` to `barge_in.py` (mirrors the identical helper in `wake_word.py`; kept separate to avoid a circular import since `wake_word` imports `barge_in`). The detected index is passed as `input_device_index` when opening the barge-in stream.

### 5. `keyboard` NameError in `Audio.on_press` (`common/audio.py` line 90)
- ✅ Fixed: store the lazily-imported `keyboard` module on `self._keyboard` inside `init_recording()`, then reference it via `getattr(self, '_keyboard', None)` in `on_press()`. This keeps the lazy import (avoids display-related import failures on headless systems) while making the symbol reachable from the callback.

**Test Results:**
- Total tests: 1056
- All passing ✅

spideyz0r:

### 1. `common/audio.py` line 35 — AUDIODEV overwritten even when user set it
- ✅ Added `"AUDIODEV" not in os.environ` guard before both assignment sites (normal path and exception fallback), so a user-supplied `AUDIODEV` is never overridden.

### 2. `README.md` line 56 — "absolute path" too strict
- ✅ Changed "absolute path to a custom \`.onnx\` file" → "path to a \`.onnx\` file" in both the prose and the config comment. The implementation accepts any string ending in \`.onnx\`, not just absolute paths.

### 3. `tests/test_wake_word.py` — local variable still named `mock_porcupine`
- ✅ Renamed `mock_porcupine` → `mock_detector` in all three `TestWakeWordModeStateIdle` tests (`test_state_idle_detects_wake_word`, `test_state_idle_plays_confirmation_beep`, `test_state_idle_handles_stream_error`).

### 4. `common/openwakeword_detector.py` — `int()` rounding and resample length mismatch
- ✅ `frame_length` property now uses `round()` instead of `int()` to avoid off-by-one drift at common device rates.
- ✅ After `resample_poly`, the output is trimmed to `[:_FRAME_LENGTH]` if longer, or zero-padded to `_FRAME_LENGTH` if shorter, guaranteeing the model always receives exactly 1280 samples regardless of rounding differences.

### Previously fixed in Round 1 — skipped

#### `audio.py` — keyboard `NameError`
- ⏭️ Fixed in Round 1 by lazy-importing `keyboard` via `pynput` inside `init_recording()` and storing it as `self._keyboard`. `on_press()` uses `getattr(self, '_keyboard', None)`. No change needed.

#### `openwakeword_detector.py` — resample output not clipped to int16 range
- ⏭️ Fixed in Round 1 by wrapping `resample_poly` output in `np.clip(..., -32768, 32767).astype(np.int16)`. No change needed.

#### `barge_in.py` — stream opened at `sample_rate` instead of `device_sample_rate`
- ⏭️ Fixed in Round 1 by opening the PyAudio stream at `detector_instance.device_sample_rate`. No change needed.

#### `barge_in.py` / `wake_word.py` — `_find_hw_input_device` duplicated
- ⏭️ Intentional duplication to avoid a circular import: `wake_word.py` imports `barge_in.py`, so a shared helper module would create a cycle. Both copies are identical and are flagged with a comment explaining the rationale. No change needed.

#### SDL device-detection block — "move to `initialize_audio()`"
- ⏭️ The block must run at **import time** to set `AUDIODEV` before `pygame` is imported (the very next statement). Moving it inside a method called later would let pygame initialise with the wrong device. Skipped.

**Test Results:**
- Total: 1056 passed, 15 subtests passed
- All passing ✅

spideyz0r:

### 1. `_find_hw_input_device()` can raise / `dev["name"]` KeyError (`wake_word.py` line 40, `barge_in.py` line 26)
- ✅ Wrapped the device-enumeration loop in `try/except Exception` in both `wake_word.py` and `barge_in.py`; returns `None` and logs a `WARNING` on any probe failure so wake-word and barge-in still start rather than crashing.
- ✅ Changed `dev["name"]` to `dev.get("name", "")` in both debug log calls for consistency with the regex guard on the same field.

### 2. `tests/test_barge_in.py` line 82 — mock missing `device_sample_rate`
- ✅ Added `mock_detector.device_sample_rate = 16000` to all 5 detection-loop test mocks so the tests validate the interface the implementation actually uses (`detector_instance.device_sample_rate` when opening the PyAudio stream).

### 3. `tests/test_wake_word.py` line 73 — `mock_config.rate` not set in setUp
- ✅ Added `self.mock_config.rate = 16000` to `TestWakeWordModeInitialize.setUp()` so `_create_detector_instance()` receives a realistic integer instead of a bare `Mock`.

### 4. `tests/test_wake_word.py` line 96 — mock missing `device_sample_rate`
- ✅ Added `mock_detector.device_sample_rate = 16000` to all 9 mock_detector instances in `test_wake_word.py` so tests reflect the full interface `WakeWordMode` relies on.

### 5. `tests/test_openwakeword_detector.py` line 24 — `test_import_succeeds` never fails
- ✅ Replaced the trivially-true `self.assertTrue(True)` with `self.assertIsNotNone(OpenWakeWordDetector)` and removed the unused `@patch` decorator. The test now actually validates that the class is importable and non-None.

### 6. `README.md` line 56 — Pi/Linux tflite-runtime install note
- ✅ Added a "Raspberry Pi / Linux install note" block with the one-liner stub install workaround, matching what the CI workflow already does.

### Skipped

#### `common/vad_recorder.py` line 138 — separate `vad_initial_silence_timeout` config
- ⏭️ Reusing `vad_silence_duration` for the pre-speech grace period is intentional: it gives the mic a reasonable idle window before giving up without introducing a second knob. Adding `vad_initial_silence_timeout` would add config complexity for a corner case. If users report false negatives from brief pauses, a dedicated config can be added then.

#### `common/audio.py` line 41 — move SDL ALSA block to `initialize_audio()`
- ⏭️ Same justification as Round 2: the block sets `AUDIODEV` immediately before `import pygame` (the very next statement in the module). Moving it to an instance method means pygame would already be imported with the wrong device. The AUDIODEV-guard concern from Round 2 has been addressed; the import-time placement is required.

**Test Results:**
- Total: 1056 passed, 15 subtests passed
- All passing ✅

spideyz0r:

### 1. `_find_hw_input_device()` duplicated in `barge_in.py` and `wake_word.py`
- ✅ Extracted into a new shared module `common/audio_devices.py`. Both `wake_word.py` and `barge_in.py` now import it from there. No circular import: the new module has no dependencies on other `common/` modules. Removed the duplication comment that was previously explaining why the copies had to be kept separate.

### 2. `common/audio.py` line 21 — ALSA errors not suppressed during SDL device enumeration
- ✅ Moved `_ALSA_NOOP_HANDLER`, `_alsa_suppression_attempted`, and `_suppress_alsa_errors()` above the SDL ALSA detection block. Added a `_suppress_alsa_errors()` call immediately after `os.environ["SDL_AUDIODRIVER"] = "alsa"` and before `pyaudio.PyAudio()`, so ALSA stderr noise is silenced during the device probe just as it is elsewhere.

### 3. `common/vad_recorder.py` line 136 — comment says "30s" but condition uses `vad_silence_duration`
- ✅ Updated the comment to accurately describe the behaviour: "hold the mic for the full `vad_timeout` window" rather than a hardcoded "30s", and clarified that the early-exit uses `vad_silence_duration` to avoid that wait.

### 4. `README.md` line 62 — tflite Python version range may drift
- ✅ Changed "no wheels for Python 3.10–3.12" → "no compatible wheels for current Python versions on Linux" to avoid future drift as Python versions change.

### Previously fixed / skipped (carried from Round 3)

All Round 3 comments re-appeared in this batch — all were already addressed. Specifically:
- `tests/test_barge_in.py`: `mock_detector.device_sample_rate` ✅ Round 3
- `common/wake_word.py` + `barge_in.py`: try/except + `dev.get("name", "")` ✅ Round 3
- `tests/test_wake_word.py`: `mock_config.rate` + `mock_detector.device_sample_rate` ✅ Round 3
- `tests/test_openwakeword_detector.py`: trivial test replaced ✅ Round 3
- `README.md`: Pi tflite note added ✅ Round 3

#### `common/vad_recorder.py` line 138 — separate `vad_initial_silence_timeout` config
- ⏭️ Same justification as Round 3: adding a second config knob for a corner case is over-engineering for now. The current behaviour is a reasonable default; a dedicated config can be added if users report real issues.

#### `common/audio.py` — move SDL block to `initialize_audio()`
- ⏭️ Same justification as Rounds 2 and 3: `AUDIODEV` must be set before `import pygame` (the next statement after the block). Moving it to an instance method would let pygame initialise with the wrong device on Linux.

**Test Results:**
- Total: 1056 passed, 15 subtests passed
- All passing ✅

spideyz0r:
## Round 5 Copilot Feedback

### `common/audio.py` line 46 — move SDL ALSA block to `initialize_audio()` — Skipped (4th time)
- ⏭️ Same justification as Rounds 2, 3, and 4: `SDL_AUDIODRIVER` and `AUDIODEV` must be set **before** `import pygame` (line 65, the very next statement after the block). pygame reads these env vars at import time to select its audio backend. Moving the block into `initialize_audio()` — called later at runtime — means pygame has already been imported with the wrong device by the time the method runs, so the detection would have no effect.

  The concern about side effects in CLI/test contexts is valid, but the fix would be to guard the `import pygame` itself (making pygame lazy), which is a much larger refactor outside the scope of this PR. The current placement is the minimum correct implementation.

  If this keeps coming up: a follow-up PR could lazily import pygame inside `play_audio_file()` and `stop_playback()` only, which would also naturally defer the SDL detection. Happy to do that separately after this PR merges.

No code changes. No new commit.

**Test Results:**
- Total: 1056 passed, 15 subtests passed
- All passing ✅

spideyz0r:

### 1. `common/openwakeword_detector.py` — docstring says "absolute path" but implementation accepts any `.onnx`
- ✅ Updated docstring: "absolute path to a custom .onnx file" → "path (absolute or relative) to a custom .onnx file", matching the actual `os.path.isabs() or model_name.endswith('.onnx')` check in the implementation.

### 2. `common/openwakeword_detector.py` — `warnings.simplefilter("ignore")` too broad
- ✅ Replaced the blanket `simplefilter("ignore")` with two targeted filters: `simplefilter("ignore", DeprecationWarning)` and `simplefilter("ignore", UserWarning)`. These cover the expected noisy warnings from openWakeWord/onnxruntime at model load time while leaving other warning categories (e.g., `RuntimeWarning`, `FutureWarning`) visible.

### 3. `tests/test_openwakeword_detector.py` — unused `MagicMock` import
- ✅ Removed `MagicMock` from the import line. Only `Mock` and `patch` are used.

### Skipped

#### `common/audio.py` — move SDL ALSA block to `initialize_audio()` (6th time)
- ⏭️ Same technical constraint as Rounds 2–5: `SDL_AUDIODRIVER` and `AUDIODEV` must be set before `import pygame` (the immediately following statement). pygame reads these at import time to select its audio backend — no method-level deferral can work without also making the pygame import lazy. That refactor is out of scope for this PR. A follow-up lazy-pygame PR could resolve this cleanly.

**Test Results:**
- Total: 1056 passed, 15 subtests passed
- All passing ✅

spideyz0r:

### 1. `BargeInDetector._sensitivity` → `_threshold` (barge_in.py line 51)
- ✅ Renamed `self._sensitivity` to `self._threshold` in `__init__` and updated `_create_detector_instance` to use `self._threshold` — consistent with `OpenWakeWordDetector` naming and the parameter name `threshold`
- Updated `tests/test_barge_in.py` assertion from `d._sensitivity` to `d._threshold`

### 2. `.onnx` path validation in `configuration.py` (line 507)
- ✅ Added `elif self.openwakeword_model.endswith(".onnx") and not os.path.exists(self.openwakeword_model): errors.append(...)` — fails fast at config validation rather than at wake-word mode startup

### 3. Add `scipy` to ImportError dependency hint (sandvoice.py line 940)
- ✅ Added `scipy` to the hint string alongside `openwakeword`, `onnxruntime`, `webrtcvad`, and `PyAudio`

### 4. SDL block location (audio.py) — Skipped
- ⏭️ Reason: Same as Rounds 5 and 6. The SDL block must run **before** `import pygame` at module level; moving it inside `initialize_audio()` would mean `import pygame` (line 71) executes before the env vars are set, breaking ALSA device selection entirely. This is an intentional architectural constraint, not an oversight.

**Test Results:**
- Total tests: 1056
- All passing ✅

spideyz0r:
## Round 8 Copilot Feedback — All Items Already Addressed

All comments raised in this review round were fixed in earlier rounds. No new code changes required.

### `common/barge_in.py` — stream rate mismatch (line 259)
- ⏭️ Already fixed. The stream is opened at `detector_instance.device_sample_rate` (not `sample_rate`) since Round 4. Current code: `rate=detector_instance.device_sample_rate`.

### `common/barge_in.py` — `_find_hw_input_device` not called / duplicated
- ⏭️ Already fixed in Round 4. `_find_hw_input_device` was extracted to `common/audio_devices.py` and is imported/called in both `wake_word.py` and `barge_in.py`. `input_device_index=input_device_index` is passed to `pa.open()`.

### `common/audio_devices.py` — `_find_hw_input_device` can raise, `dev["name"]` KeyError
- ⏭️ Already handled. The entire enumeration is wrapped in `try/except Exception` (returns `None` and logs a warning on failure). The code uses `dev.get("name", "")` — no KeyError possible.

### `common/audio.py` — keyboard `NameError` in `on_press()`
- ⏭️ Already fixed. `init_recording()` stores `keyboard` on `self._keyboard`; `on_press()` retrieves it via `getattr(self, '_keyboard', None)` and uses `_kb.Key.esc`. No `NameError` possible.

### `common/audio.py` — `AUDIODEV` overwritten even when user sets it
- ⏭️ Already guarded. The code has `if "AUDIODEV" not in os.environ:` before setting `AUDIODEV`.

### `common/audio.py` — SDL ALSA detection at module import time (recurring)
- ⏭️ Same skip as Rounds 5, 6, and 7. The SDL block **must** run before `import pygame` at module level. Moving it into `initialize_audio()` would execute `import pygame` (line 71) before the env vars are set, breaking ALSA device selection entirely.

### `common/vad_recorder.py` — comment mentions "30s"
- ⏭️ Already fixed in Round 4. The comment now reads: "hold the mic for the full vad_timeout window" — no hard-coded "30s".

### `common/vad_recorder.py` — separate `vad_initial_silence_timeout` config
- ⏭️ Over-engineering skip. The current `vad_silence_duration` bailout works correctly for the no-speech case. Adding a separate config key for initial vs. post-speech silence would add complexity without a clear user-facing benefit.

### `common/openwakeword_detector.py` — resample clipping, `int()` truncation
- ⏭️ Both fixed in Round 6. Resampled output is clipped to `[-32768, 32767]` before `.astype(np.int16)`. `frame_length` uses `round()` instead of `int()`. Output is also trimmed/padded to exactly `_FRAME_LENGTH`.

### `tests/test_barge_in.py` — `mock_detector.device_sample_rate` missing
- ⏭️ Already set. All detection-loop test mocks include `mock_detector.device_sample_rate = 16000` (lines 83, 372, 402, 448, 474).

### `tests/test_wake_word.py` — `mock_detector.device_sample_rate` missing, `mock_porcupine` naming, `mock_config.rate` missing
- ⏭️ All fixed in Rounds 2–3. All 9 `mock_detector` instances have `mock_detector.device_sample_rate = 16000`. No `mock_porcupine` references remain. `self.mock_config.rate = 16000` set in `setUp()`.

### `tests/test_openwakeword_detector.py` — `test_import_succeeds` unconditionally asserts `True`
- ⏭️ Fixed in Round 6. Now uses `self.assertIsNotNone(OpenWakeWordDetector)`.

### `README.md` — absolute path wording, Pi/Linux install note, Python version range
- ⏭️ All fixed in Round 3. Path wording generalized to "path to a `.onnx` file"; Pi/Linux tflite stub note added; Python version range generalized.

**Test Results:**
- Total tests: 1056
- All passing ✅

spideyz0r:
## Round 9 Copilot Feedback Addressed

### 1. `common/audio_devices.py` — `_find_hw_input_device` has private-convention underscore but is used publicly
- ✅ Renamed `_find_hw_input_device` → `find_hw_input_device`; updated imports and call sites in `common/wake_word.py` and `common/barge_in.py`

### 2. `README.md` / `docs/CUSTOM_WAKE_WORDS.md` — still references Porcupine keys
- ✅ Rewrote `docs/CUSTOM_WAKE_WORDS.md` entirely for openWakeWord: removed all Porcupine references (`porcupine_access_key`, `porcupine_keyword_paths`, `.ppn` files, Picovoice Console). New content covers built-in models, custom ONNX training via Colab, config reference, and Pi install notes.

### 3. `common/openwakeword_detector.py` — duplicate INFO log on detection
- ✅ Downgraded `logger.info("Wake word detected: ...")` → `logger.debug(...)`. Both `WakeWordMode` (line 288 of `wake_word.py`) and `BargeInDetector` (line 274 of `barge_in.py`) already log at INFO when detection occurs — the detector-level log was redundant.

### 4. `common/vad_recorder.py` — `vad_silence_duration` reused for initial-silence timeout — Skipped
- ⏭️ Introducing a separate `vad_initial_silence_timeout` config key would add complexity for a feature that is not yet requested. The current behavior (discard if no speech within `vad_silence_duration`) is intentional and documented. A dedicated config can be added in a follow-up plan if users report the current window is too short.

**Test Results:**
- Total tests: 1056
- All passing ✅

spideyz0r:
## Round 10 Copilot Feedback Addressed

### 1. `common/configuration.py` — `os.path.exists` doesn't expand `~` or env vars
- ✅ Added `expanded = os.path.expandvars(os.path.expanduser(self.openwakeword_model))` before the existence check. Validation now runs against the expanded path. If the file exists, `self.openwakeword_model` is updated to the expanded path so downstream code (OpenWakeWordDetector init) opens the correct absolute path.

### 2. `common/audio.py` — SDL probe runs before `_suppress_alsa_errors()` — Skipped
- ⏭️ Calling `_suppress_alsa_errors()` inside the SDL probe block was tried in Round 4 and caused a CI segfault (exit code 139): `snd_lib_error_set_handler()` called at import time on Linux CI runners without real ALSA hardware triggered a PortAudio crash. The call was removed in the Round 4 fixup commit (`a691d84`). Moving the entire block into `initialize_audio()` is also not viable because `import pygame` (line 71) executes at module level before any method runs — env vars must be set before that import or SDL ignores them. No safe fix exists without a larger refactor that goes beyond this PR's scope.

**Test Results:**
- Total tests: 1056
- All passing ✅

spideyz0r:

### 1. `common/configuration.py` — case-sensitive `.onnx` check
- ✅ Changed `self.openwakeword_model.endswith(".onnx")` → `self.openwakeword_model.lower().endswith(".onnx")` so paths like `MODEL.ONNX` are handled correctly.

### 2. `common/openwakeword_detector.py` — case-sensitive `.onnx` check
- ✅ Changed `model_name.endswith(".onnx")` → `model_name.lower().endswith(".onnx")` for the same reason.

### 3. `README.md` — `wake_word_sensitivity` comment says "higher = more sensitive" (wrong)
- ✅ Updated to `# 0.0-1.0; higher = stricter (fewer false positives), lower = more sensitive` — accurately describes the threshold semantics.

### 4. `common/audio.py` — SDL probe at import time before `_suppress_alsa_errors()` — Skipped
- ⏭️ Same skip as Rounds 5–11. The SDL env vars must be set before `import pygame` (line 71), which runs at module level. Calling `_suppress_alsa_errors()` inside the SDL probe block was tried in Round 4 and caused a CI segfault (exit 139) on Linux runners. No safe in-scope fix exists without a larger refactor.

**Test Results:**
- Total tests: 1056
- All passing ✅

spideyz0r:

### 1. `common/audio.py` — `keyboard.Listener` held only in local variable, may be GC'd
- ✅ Stored the listener on `self._listener`; wrapped `start_recording()` / `convert_to_mp3()` in a `try/finally` so `self._listener.stop()` is always called when recording finishes, even on error.

### 2. `common/audio.py` — SDL probe at import time before `_suppress_alsa_errors()` — Skipped
- ⏭️ Recurring skip (Rounds 5–12). The SDL env vars must be set before `import pygame` at module level. Calling `_suppress_alsa_errors()` inside the SDL probe block was attempted in Round 4 and caused a CI segfault (exit 139) on Linux runners. No safe in-scope fix exists without a larger refactor.

**Test Results:**
- Total tests: 1056
- All passing ✅

spideyz0r:

### 1. `common/configuration.py` — `openwakeword_model` not stripped before `.endswith` check
- ✅ Added `self.openwakeword_model = self.openwakeword_model.strip()` in the `else` branch before the `.onnx` check. This ensures leading/trailing whitespace never causes the path-expansion branch to be silently skipped, and the stripped value flows into `OpenWakeWordDetector` downstream.

### 2. `common/audio.py` — SDL probe without `_suppress_alsa_errors()` — Skipped
- ⏭️ Recurring skip (Rounds 5–13). Calling `_suppress_alsa_errors()` inside the SDL probe block was attempted in Round 4 and caused a CI segfault (exit 139). Moving the block into `initialize_audio()` is not viable because `import pygame` runs at module level before any method. No safe in-scope fix.

**Test Results:**
- Total tests: 1056
- All passing ✅

PR #151, opened 2026-05-09
Fix deprecated wakeword_model_paths kwarg

spideyz0r:
## Summary
- Replace `wakeword_model_paths=` with `wakeword_models=` in the `Model()` call in `common/openwakeword_detector.py`

## Problem
Every startup printed:
```
WARNING root: DEPRECATION: keyword argument 'wakeword_model_paths' is no longer valid and will be removed in future releases. Use 'wakeword_models' instead.
```

The deprecated kwarg was used for `.onnx` file path models; built-in named models already used `wakeword_models`. Both branches now use the current API.

## Changes
- `common/openwakeword_detector.py`: one-line kwarg rename

### DIFF
--- a/common/openwakeword_detector.py
+++ b/common/openwakeword_detector.py
@@ -45,15 +45,15 @@ class OpenWakeWordDetector:
             self._resample_up = None
             self._resample_down = None
 
+        if os.path.isabs(model_name) or model_name.lower().endswith(".onnx"):
+            self._prediction_key = os.path.splitext(os.path.basename(model_name))[0]
+        else:
+            self._prediction_key = model_name
+
         with warnings.catch_warnings():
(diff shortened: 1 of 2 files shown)


spideyz0r:

### 1. Test still asserting `wakeword_model_paths` (openwakeword_detector.py:52) — Stale
- ⏭️ Already fixed in the previous commit: test renamed to `test_init_onnx_path_uses_wakeword_models` and assertion updated to `wakeword_models=`.

### 2. `if/else` duplicates identical `Model(...)` call (openwakeword_detector.py:55)
- ✅ Hoisted `Model(wakeword_models=[model_name], inference_framework="onnx")` out of the conditional. The branch now only sets `_prediction_key`, which is the only thing that actually differs.

**Test Results:** 1059 passed ✅

PR #153, opened 2026-05-10
Plan 59: Drop Python <3.9 zoneinfo fallback

spideyz0r:
## Summary
- Replace `try/except ImportError` around `from zoneinfo import ZoneInfo` with a direct import in `plugins/weather/plugin.py`, `plugins/greeting/plugin.py`, and `common/scheduler.py`
- Remove the `ZoneInfo is None` guard from `_resolve_tz()` in all three files

`zoneinfo` is in the stdlib since Python 3.9 (Oct 2020). Both target platforms exceed this: macOS M1 runs 3.14, Raspberry Pi OS Bullseye ships 3.9+. The fallback was dead code that silently degraded — if triggered, `ZoneInfo = None` would break cache key rotation and timezone display without any error.

## Test plan
- [x] 187 weather, greeting, and scheduler tests pass
- [x] No `ZoneInfo is None` code paths remain in any file

### DIFF
--- a/common/scheduler.py
+++ b/common/scheduler.py
@@ -4,10 +4,7 @@ import threading
 from datetime import datetime, timedelta, timezone
 from typing import Any, Callable, Iterable, Optional
 
-try:
-    from zoneinfo import ZoneInfo
-except ImportError:
-    ZoneInfo = None  # Python < 3.9 fallback; times will print in UTC
+from zoneinfo import ZoneInfo
 
 from common.db import SchedulerDB, ScheduledTask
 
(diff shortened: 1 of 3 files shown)


spideyz0r:

### 1. `ZoneInfo is None` path remains in `common/scheduler.py` (weather/plugin.py line 8)
- ✅ Applied the same fix to `common/scheduler.py`: replaced the `try/except ImportError` block with a direct `from zoneinfo import ZoneInfo` import
- ✅ Removed the `if ZoneInfo is None:` guard from `TaskScheduler._resolve_tz()`
- The PR description claim "no `ZoneInfo is None` code paths remain" is now accurate

**Test Results:**
- 187 tests passing ✅ (weather, greeting, scheduler)

spideyz0r:

### 1. PR description doesn't mention scheduler change (`common/scheduler.py` line 8)
- ✅ Updated the PR description to include `common/scheduler.py` in the Summary and Test plan sections

No code changes — the scheduler fix was already in the branch from round 1.

**Test Results:**
- 187 tests passing ✅

